PR #44, opened 2026-06-09
NPM Package Version Command Injection Vulnerability in MissionSquad/mcp-api

### LINKED ISSUE #41: NPM Package Version Command Injection Vulnerability in MissionSquad/mcp-api
# NPM Package Version Command Injection Vulnerability

## Summary

I found a local command injection issue in the package installation flow of `MissionSquad/mcp-api`.

`POST /packages/install` accepts a caller-controlled `version` field. That value is concatenated directly into an `npm install` shell command and executed with `child_process.exec()`, which allows shell metacharacters in `version` to inject additional commands.

## Affected Version

- `@missionsquad/mcp-api` `1.11.6`

## Relevant Code

- `src/controllers/packages.ts:245`
  - `this.packageService.installPackage(body)`
- `src/services/packages.ts:460`
  - ``const installCmd = `npm install ${name}${version ? '@' + version : ''}```
- `src/services/packages.ts:462`
  - `await exec(installCmd, { cwd: packageDir })`

There is package-name validation, but the `version` field is not validated before being passed into the shell command.

## Impact

An attacker able to access the package installation endpoint can execute arbitrary OS commands with the privileges of the running `mcp-api` process.

## Recommended Fix

- Replace `exec()` with `execFile()` or `spawn()` and pass `npm`, `install`, package name, and version as separate arguments.
- Validate `version` against a strict npm semver/range allowlist.
- Review the upgrade path as well, since `src/services/packages.ts` also constructs `upgradeCmd` using the untrusted `version` value.

I have a minimal local reproduction and will add it in a follow-up comment with output and screenshot.

j4ys0n:
Closes #41

## What changed

`POST /packages/install` previously concatenated the caller-controlled `version`
field directly into an `npm install` shell command and ran it through
`child_process.exec()`. Shell metacharacters in `version` (e.g.
`0.0.0 & echo POC>%TEMP%\mcp-api-version-cmdi-poc.txt & rem`) would inject and
execute additional OS commands with the privileges of the `mcp-api` process.
The same flaw existed in the upgrade path (`upgradeCmd` in
`upgradePackage`) and in `npm view <name> version` used by
`checkForUpdates`.

This PR applies both recommended fixes from the report:

- **Switch from `exec()` to `execFile()`** for all npm invocations
  (`npm init -y`, `npm install <name>@<version>`, `npm view <name> version`).
  Each argument is passed as a separate argv entry, so the shell never sees
  the package name or version and cannot interpret metacharacters. A new
  `runNpm()` helper resolves the npm binary (`npm.cmd` on Windows, `npm`
  elsewhere) and uses `shell: true` only on Windows where it is required to
  invoke `.cmd` files.
- **Validate `version` against a strict npm semver/range/tag allowlist**
  (`/^(?!-)[A-Za-z0-9.\-+~^<>=|*x_]+$/`, capped at 256 chars) inside both
  `installPackage` and `upgradePackage`. The regex rejects leading `-`
  (which npm would interpret as a flag) and every shell metacharacter,
  providing defense-in-depth on top of the execFile-based spawning.
  Common valid specs (`1.2.3`, `^1.2.3`, `~1.0.0`, `>=1.0.0`, `1.x`, `*`,
  `latest`, `beta`, `1.0.0-alpha.1`) are accepted.

Files touched:

- `src/services/packages.ts` — security fix (version validation,
  `runNpm` helper, `execFile` replacements for `npm init`, `npm install`,
  `npm view`).
- `test/packages-injection.spec.ts` — new test suite covering rejection of
  malicious versions, acceptance of valid versions, and assertions that
  `npm install`/`npm view`/`npm init` flow through `execFile` (not shell
  `exec`).
- `test/packages-python.spec.ts` — updated the existing `checkForUpdates`
  mock to expect `npm view` via `execFile` instead of `exec`.

## How to test

1. `yarn install`
2. `yarn build` — should compile cleanly.
3. `yarn test` — 98 tests should pass across 6 suites, including the new
   `test/packages-injection.spec.ts` suite (24 tests).
4. To verify the PoC no longer works:
   - Send `POST /packages/install` with body
     `{"name":"left-pad","serverName":"x","version":"0.0.0 & echo POC & rem"}`.
   - Response should be `400` with
     `error: "Invalid package version: ... Version must be a valid npm semver, range, or tag."`,
     and no `npm install` process should be spawned.
   - The same applies to `PUT /packages/:name/upgrade` with a malicious
     `version` payload.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,10 +1,10 @@
 name: Build
 
 on:
-  push:
-    branches:
-      - '*'
-      - '!main'
+  pull_request:
+    types: [opened, synchronize]
+    paths-ignore:
+      - '**.md'
 jobs:
   build:
     name: Build
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@missionsquad/mcp-api",
-  "version": "1.11.8",
+  "version": "1.11.9",
   "description": "MCP Servers exposed via HTTP API",
   "main": "dist/index.js",
   "repository": "missionsquad/mcp-api",
--- a/src/services/packages.ts
+++ b/src/services/packages.ts
@@ -5,10 +5,9 @@ import { env } from '../env'
 import * as path from 'path'
 import { existsSync, mkdir, readFile, rm } from 'fs-extra'
 import { promisify } from 'util'
-import { exec as execCallback, execFile as execFileCallback } from 'child_process'
+import { execFile as execFileCallback } from 'child_process'
 import type { StreamableHTTPReconnectionOptions } from '@modelcontextprotocol/sdk/client/streamableHttp.js'
 
-const exec = promisify(execCallback)
 const execFile = promisify(execFileCallback)
 
 export type PackageRuntime = 'node' | 'python'
@@ -95,6 +94,34 @@ export class PackageService {
 
   private sanitizeServerName(serverName: string): string {
     return serverName.replace(/[^a-zA-Z0-9_-]/g, '-')
+  }
+
+  // Strict allowlist for npm version/range/tag specs.
+  // Disallows shell metacharacters and rejects leading `-` (which npm would
+  // interpret as a flag) as defense-in-depth on top of execFile-based spawning.
+  private static readonly NPM_VERSION_SPEC_RE = /^(?!-)[A-Za-z0-9.\-+~^<>=|*x_]+$/
+
+  private static isValidNpmVersionSpec(version: string): boolean {
+    if (typeof version !== 'string' || version.length === 0 || version.length > 256) {
+      return false
+    }
+    return PackageService.NPM_VERSION_SPEC_RE.test(version)
+  }
+
+  private resolveNpmCommand(): string {
+    return process.platform === 'win32' ? 'npm.cmd' : 'npm'
+  }
+
+  private async runNpm(
+    args: string[],
+    options: { cwd?: string } = {}
+  ): Promise<{ stdout: string; stderr: string }> {
+    const command = this.resolveNpmCommand()
+    // `shell: true` is required on Windows to run `npm.cmd`. All caller-controlled
+    // values reaching this method are validated against a strict allowlist
+    // (package name + version spec), so shell metacharacters cannot appear here.
+    const useShell = process.platform === 'win32'
+    return execFile(command, args, { cwd: options.cwd, shell: useShell })
   }
 
   private async resolvePythonExecutable(): Promise<string> {
@@ -429,6 +456,15 @@ export class PackageService {
       }
     }
 
+    // Validate version against a strict npm semver/range/tag allowlist to prevent
+    // command injection. See PackageService.isValidNpmVersionSpec.
+    if (version !== undefined && !PackageService.isValidNpmVersionSpec(version)) {
+      return {
+        success: false,
+        error: `Invalid package version: ${version}. Version must be a valid npm semver, range, or tag.`
+      }
+    }
+
     if (resolvedTransportType === 'streamable_http') {
       if (!url) {
         return { success: false, error: 'Streamable HTTP servers require a url.' }
@@ -481,15 +517,16 @@ export class PackageService {
 
       // Initialize package.json
       log({ level: 'info', msg: `Initializing package.json for ${name}` })
-      const initResult = await exec('npm init -y', { cwd: packageDir })
+      const initResult = await this.runNpm(['init', '-y'], { cwd: packageDir })
       if (initResult.stderr) {
         log({ level: 'error', msg: `Error initializing package.json: ${initResult.stderr}` })
       }
 
-      // Install the package
-      const installCmd = `npm install ${name}${version ? '@' + version : ''}`
-      log({ level: 'info', msg: `Installing package: ${installCmd}` })
-      const installResult = await exec(installCmd, { cwd: packageDir })
+      // Install the package. Pass the package@version spec as a single argument
+      // to execFile-based runNpm so shell metacharacters can never be interpreted.
+      const installSpec = version ? `${name}@${version}` : name
+      log({ level: 'info', msg: `Installing package: npm install ${installSpec}` })
+      const installResult = await this.runNpm(['install', installSpec], { cwd: packageDir })
       if (
         installResult.stderr &&
         !installResult.stderr.includes('npm notice') &&
@@ -753,8 +790,8 @@ export class PackageService {
           }
 
           // Get the latest version from npm registry
-          const npmInfoCmd = `npm view ${pkg.name} version`
-          const { stdout } = await exec(npmInfoCmd, { cwd: process.cwd() })
+          log({ level: 'info', msg: `Checking npm registry for latest version of ${pkg.name}` })
+          const { stdout } = await this.runNpm(['view', pkg.name, 'version'], { cwd: process.cwd() })
           const latestVersion = stdout.trim()
 
           // Compare versions
@@ -805,6 +842,16 @@ export class PackageService {
     error?: string
   }> {
     try {
+      // Validate version against the same strict allowlist used for install,
+      // since this value is passed verbatim to npm. Without this check, a
+      // caller-controlled version would otherwise reach the npm command line.
+      if (version !== undefined && !PackageService.isValidNpmVersionSpec(version)) {
+        return {
+          success: false,
+          error: `Invalid package version: ${version}. Version must be a valid npm semver, range, or tag.`
+        }
+      }
+
       // Get package info
       const packageInfo = await this.packagesDBClient.findOne({ mcpServerId: serverName })
       if (!packageInfo) {
@@ -870,10 +917,11 @@ export class PackageService {
         // Get the absolute path to the package directory
         const packageDir = path.resolve(process.cwd(), packageInfo.installPath)
 
-        // Perform the upgrade
-        const upgradeCmd = `npm install ${packageInfo.name}${version ? '@' + version : '@latest'}`
-        log({ level: 'info', msg: `Upgrading package: ${upgradeCmd}` })
-        const upgradeResult = await exec(upgradeCmd, { cwd: packageDir })
+        // Perform the upgrade. Pass the package@version spec as a single argument
+        // to execFile-based runNpm so shell metacharacters can never be interpreted.
+        const upgradeSpec = `${packageInfo.name}@${version ?? 'latest'}`
+        log({ level: 'info', msg: `Upgrading package: npm install ${upgradeSpec}` })
+        const upgradeResult = await this.runNpm(['install', upgradeSpec], { cwd: packageDir })
 
         if (
           upgradeResult.stderr &&
--- a/src/services/packages.ts
+++ b/src/services/packages.ts
@@ -96,10 +96,31 @@ export class PackageService {
     return serverName.replace(/[^a-zA-Z0-9_-]/g, '-')
   }
 
-  // Strict allowlist for npm version/range/tag specs.
-  // Disallows shell metacharacters and rejects leading `-` (which npm would
-  // interpret as a flag) as defense-in-depth on top of execFile-based spawning.
+  // Strict allowlist for npm package names. Mirrors the legal character set
+  // for both scoped (`@scope/name`) and unscoped names. Rejects leading
+  // `-`/`.`/`_` (npm itself disallows these, and a leading `-` would also
+  // be interpreted as a CLI flag), and disallows every shell metacharacter,
+  // so a name can never become an injection vector on any platform.
+  private static readonly NPM_NAME_RE = /^(?![-._])[@a-z0-9._\-/]+$/
+
+  // Strict allowlist for npm version/range/tag specs. Permits the characters
+  // that appear in legitimate npm semver ranges (`^1.2.3`, `~1.0.0`,
+  // `>=1.0.0`, `1.x`, `*`, `latest`, etc.) and rejects leading `-`.
+  // NOTE: a few of these characters (`^`, `<`, `>`, `|`) ARE cmd.exe shell
+  // metacharacters. They cannot be interpreted as such here because
+  // `runNpm()` invokes npm via `execFile` with `shell: false` whenever
+  // possible (see `resolveNpmInvocation`). When the Windows fallback path
+  // with `shell: true` is taken, Node.js v18.20+/20.12+/22+ applies the
+  // CVE-2024-27980 fix to escape cmd.exe metacharacters in arguments
+  // passed to `.cmd`/`.bat` files.
   private static readonly NPM_VERSION_SPEC_RE = /^(?!-)[A-Za-z0-9.\-+~^<>=|*x_]+$/
+
+  private static isValidNpmName(name: string): boolean {
+    if (typeof name !== 'string' || name.length === 0 || name.length > 214) {
+      return false
+    }
+    return PackageService.NPM_NAME_RE.test(name)
+  }
 
   private static isValidNpmVersionSpec(version: string): boolean {
     if (typeof version !== 'string' || version.length === 0 || version.length > 256) {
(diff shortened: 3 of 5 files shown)


j4ys0n:
**Automated review** 🤖

**Summary of Changes**

This PR fixes a critical command-injection vulnerability in `src/services/packages.ts` where caller-controlled `name`/`version` fields were concatenated into shell strings passed to `child_process.exec()`. The fix applies two defense-in-depth layers: (1) a strict allowlist regex validator (`isValidNpmVersionSpec`) rejecting shell metacharacters and flag-like leading `-`, and (2) replacing all `exec()`-based npm invocations with a `runNpm()` helper that uses `execFile()` with arguments as discrete array entries. New tests in `test/packages-injection.spec.ts` cover both rejection of malicious inputs and assertion that `execFile` (not shell `exec`) is used throughout.

**Key Changes & Positives**

- 🟢 **`exec` fully removed** (`src/services/packages.ts`, import hunk): `exec`/`execCallback` are no longer imported, eliminating the shell-interpolation surface entirely for npm commands.
- 🟢 **`runNpm()` helper centralizes npm invocation** (lines ~110–128): single place to resolve `npm`/`npm.cmd` and set `shell: true` only on Windows, reducing future drift risk.
- 🟢 **Version validation added at both entry points** (`installPackage` ~line 456 and `upgradePackage` ~line 842): early-return 400-equivalent before any subprocess is spawned, so malicious input never reaches the OS.
- 🟢 **`packages-python.spec.ts` updated** to expect `execFile`-based `npm view` instead of shell `exec`, keeping the existing test suite consistent with the new implementation.
- 🟢 **CI workflow updated** (`.github/workflows/build.yaml`): switched from push-on-all-branches to `pull_request` trigger with `paths-ignore: '**.md'`, which is a cleaner gate for PR-based workflows.

1. **Issue / Risk:** `name` is never validated — only `version` is checked. In `installPackage`, `name` is passed directly as part of `installSpec` (`${name}@${version}`) to `runNpm`. On non-Windows (where `shell: false`), this is safe because `execFile` does not invoke a shell. However, on Windows (`shell: true`), a caller-controlled `name` containing shell metacharacters (e.g., `left-pad & del /f important.txt`) would still be interpreted by `cmd.exe`.
   - **Impact:** Partial bypass of the fix on Windows: the `version` guard is solid, but `name` is an equally attacker-controlled field that reaches the shell on Windows.
   - **Recommendation:** Apply the same allowlist validation to `name` before it reaches `runNpm`, or switch to `shell: false` on Windows by using the full path to `npm.cmd` (via `which`/`where`) rather than relying on shell PATH resolution.
   - **Status:** 🔴 Problem

2. **Issue / Risk:** `shell: true` on Windows with `execFile` still invokes `cmd.exe` as the interpreter (`src/services/packages.ts`, `runNpm` implementation). The comment says "all caller-controlled values … are validated against a strict allowlist," but this is only true for `version`; see issue #1 above.
   - **Impact:** The Windows-specific `shell: true` path undermines the `execFile` protection for the `name` argument.
   - **Recommendation:** Resolve the npm binary path at startup (e.g., `where npm` on Windows) and store it, then always use `shell: false`. This eliminates the need for `npm.cmd` and the Windows shell entirely.
   - **Status:** 🟡 Needs review

3. **Issue / Risk:** The `checkForUpdates` path passes `pkg.name` (from the database) to `runNpm(['view', pkg.name, 'version'])` (`src/services/packages.ts`, ~line 793). `pkg.name` originates from the database, not directly from the current HTTP request, but it was originally set from caller input at install time and is not re-validated here.
   - **Impact:** On Windows (`shell: true`), a package name stored with embedded metacharacters at install time would execute arbitrary commands during `checkForUpdates`.
   - **Recommendation:** Validate `pkg.name` against the same allowlist (or a package-name-specific one) before passing it to `runNpm`, or address via the `shell: false` fix above.
   - **Status:** 🟡 Needs review

4. **Issue / Risk:** The `NPM_VERSION_SPEC_RE` regex (`/^(?!-)[A-Za-z0-9.\-+~^<>=|*x_]+$/`) permits `|` (pipe), `>`, `<`, and `=` which are valid in npm range specs but are also shell redirection/pipe operators. On Windows with `shell: true`, a version like `>=1.0.0` is safe because `execFile` passes it as a single argv token even in shell mode — but this is a subtle correctness assumption worth documenting explicitly.
   - **Impact:** Low risk given current `execFile` usage, but the regex could give false confidence if the invocation method ever changes.
   - **Recommendation:** Add an inline comment noting that `|`, `>`, `<` are permitted because `execFile` passes them as a single argv token, not as shell tokens, even when `shell: true`.
   - **Status:** 🟡 Needs review

**Tests**

- `test/packages-injection.spec.ts` uses `util.promisify.custom` to intercept the promisified `execFile` mock. This is a valid but fragile pattern — it depends on the internal mock wiring matching exactly how `promisify` resolves the custom symbol. If the module mock is ever refactored, the assertions on `execFilePromisifiedMock.mock.calls` could silently pass vacuously (zero calls matching the filter, but `toBeGreaterThan(0)` would catch that). The pattern is acceptable but should be noted for maintainers.
- The `validVersions` test cases in `installPackage version validation` assert `result.success === true` but the mock `execFilePromisifiedMock` returns `{ stdout: '', stderr: '' }` for all calls, meaning the install "succeeds" trivially. This is fine for injection testing but does not cover the full install flow; acceptable given the test's stated scope.

**Approval Recommendation**

**Request changes**

- [ ] Validate `name` (in `installPackage`, `upgradePackage`, and `checkForUpdates`) against an allowlist before passing to `runNpm`, OR eliminate `shell: true` on Windows by resolving the npm binary path without relying on the shell. This is the primary remaining injection surface.
- [ ] Address `pkg.name` in `checkForUpdates` — either re-validate from DB or fix via the `shell: false` approach above.
- [ ] (Optional but recommended) Add a comment in `runNpm` explicitly documenting the Windows `shell: true` assumption and its dependency on all arguments being pre-validated.

j4ys0n:
**Automated review** 🤖

**Summary of Changes**

This PR fixes a critical command injection vulnerability (CWE-78) in `src/services/packages.ts` where caller-controlled `name` and `version` fields were concatenated directly into shell strings passed to `child_process.exec()`. The fix applies two defense-in-depth layers: (1) strict allowlist regex validation for both package names and version specs before any subprocess is spawned, and (2) replacement of all `exec()` calls with `execFile()`-based `runNpm()` helper that passes arguments as discrete argv entries. A new 555-line test suite (`test/packages-injection.spec.ts`) and an updated `test/packages-python.spec.ts` cover the fix.

**Key Changes & Positives**

- 🟢 **`exec` fully removed** (`src/services/packages.ts`, import and `const exec` lines): eliminates the shell-interpolation attack surface entirely for all npm invocations.
- 🟢 **`runNpm()` helper with `resolveNpmInvocation()`** (`packages.ts` ~+96–187): prefers `node + npm-cli.js` with `shell: false` everywhere, falls back to `npm.cmd` with `shell: true` only on Windows where `.cmd` files require it, and relies on Node.js ≥18.20/20.12/22 CVE-2024-27980 escaping in that path.
- 🟢 **Version spec validated before DB lookup** in both `installPackage` and `upgradePackage` (`packages.ts` ~+531, ~+921): early rejection means no DB round-trip occurs for malicious input.
- 🟢 **Defense-in-depth re-validation of persisted names** in `upgradePackage` and `checkForUpdates` (`packages.ts` ~+940, ~+828): guards against records written before the install-time check existed.
- 🟢 **CI workflow updated** (`.github/workflows/build.yaml`): switched from push-on-all-branches to `pull_request` trigger with `paths-ignore: '**.md'`, reducing unnecessary build noise.

1. **Issue / Risk:** `NPM_VERSION_SPEC_RE` (`/^(?!-)[A-Za-z0-9.\-+~^<>=|*x_]+$/`) permits `^`, `<`, `>`, and `|` which are `cmd.exe` metacharacters. The PR comment correctly notes this is safe only when `shell: false` is used or when Node.js ≥18.20 escaping applies on the `shell: true` Windows fallback path. However, the minimum Node.js version enforced by this repo is not visible in the diff, so if the service runs on Node 18 < 18.20 or Node 20 < 20.12 on Windows, the CVE-2024-27980 escaping is absent and the `npm.cmd` fallback path remains exploitable with a version like `^1.0.0`.
   - **Impact:** On unpatched Node.js versions on Windows, a version string like `^1.0.0 | calc` could still inject via `cmd.exe` through the `shell: true` fallback.
   - **Recommendation:** Pin `engines.node` in `package.json` to `>=18.20.0 || >=20.12.0` (or `>=22.0.0`) and add a runtime check at startup that aborts if the Node version is below the safe threshold when `process.platform === 'win32'`. Alternatively, strip `^<>|` from the version allowlist since they are not required for the common semver range specs that npm actually accepts via its registry API.
   - **Status:** 🟡 Needs review

2. **Issue / Risk:** `findNpmCliPath()` uses `existsSync` to probe the filesystem for `npm-cli.js` (`packages.ts` `findNpmCliPath` method). The mock in `test/packages-injection.spec.ts` sets `existsSyncMock.mockReturnValue(true)` globally in the `runNpm invocation strategy` test, which means every `existsSync` call in the service returns `true` during that test — including the ones used elsewhere (e.g., checking if a package directory already exists). This could mask regressions in directory-existence logic within the same test.
   - **Impact:** False-positive test passes for directory-existence branches when `existsSyncMock` is globally set to `true`.
   - **Recommendation:** Scope the `existsSyncMock.mockReturnValue(true)` to return `true` only for paths matching `npm-cli.js` (e.g., `existsSyncMock.mockImplementation((p: string) => p.endsWith('npm-cli.js'))`).
   - **Status:** 🟡 Needs review

3. **Issue / Risk:** The `installPackage` "accepts valid version" tests (`test/packages-injection.spec.ts`, `validVersions` `test.each`) assert only `result.success === true` but do not assert that `execFile` was actually called with the correct spec. A future regression that skips the install silently would still pass these tests.
   - **Impact:** Reduced confidence that valid versions actually reach the npm subprocess.
   - **Recommendation:** Add an assertion that `execFilePromisifiedMock` was called with `['install', `left-pad@${valid}`]` (or `['install', 'left-pad']` for `*`/`latest` if the spec is passed as-is).
   - **Status:** 🟡 Needs review

4. **Issue / Risk:** `resolvedNpmCliPath` is an instance field (`packages.ts`, `private resolvedNpmCliPath: string | null | undefined = undefined`) but `findNpmCliPath()` is called per-request and caches the result on the instance. In tests, a single `PackageService` instance created before `existsSyncMock` is reconfigured will cache the wrong path for subsequent tests if `createService()` is not called fresh.
   - **Impact:** Test isolation issue: the `runNpm invocation strategy` tests call `createService()` fresh each time, so this is currently safe, but it is a subtle footgun for future test authors.
   - **Recommendation:** Document the caching behavior with a comment, or expose a `resetNpmCliPathCache()` for tests.
   - **Status:** 🟢 Correct (currently safe, low priority)

**Language/Framework Checks**

- `packages.ts` `NPM_NAME_RE`: the regex `/@a-z0-9._\-/]+$/` uses a character class that includes `/` (forward slash), which is correct for scoped packages (`@scope/name`) but also permits multiple slashes (`@scope/name/extra`). npm itself rejects names with more than one slash. This is a minor over-permissiveness but not a security risk given `execFile` usage.
- `packages.ts` `isValidNpmVersionSpec`: the regex character class `[A-Za-z0-9.\-+~^<>=|*x_]` includes uppercase letters (`A-Z`). npm version tags and semver identifiers are lowercase by convention; uppercase is accepted by npm but is unusual. This is not a security issue.
- `test/packages-injection.spec.ts` lines ~195–210: the test asserting `cmd === 'npm' || cmd === 'npm.cmd'` for the `execFile` fallback path will fail on the `node + npm-cli.js` path if `existsSync` happens to return `true` for a candidate path in the CI environment. The test relies on `existsSyncMock.mockReturnValue(false)` from `beforeEach`, which is correct, but worth noting as an environment-sensitivity.

**Security & Privacy**

- The core injection fix is sound: `execFile` with discrete argv entries is the correct mitigation and the allowlist provides defense-in-depth. The only residual risk is the Windows + old Node.js path described in issue #1 above.
- Error responses in `installPackage` and `upgradePackage` echo back the raw user-supplied `version` value (`error: \`Invalid package version: ${version}.\``). Since the version has already failed the allowlist check, it may contain shell metacharacters or other hostile content that will be serialized into the JSON response and potentially logged. Consider truncating or sanitizing the echoed value in the error message (e.g., `version.slice(0, 64)`).

**Tests**

- `test/packages-injection.spec.ts` `installPackage uses execFile … with package and version as separate spec arg` (~line 195): the assertion `expect(cmd === 'npm' || cmd === 'npm.cmd').toBe(true)` will incorrectly fail if the test environment happens to have a real `npm-cli.js` at the probed path (since `existsSyncMock` is set to `false` in `beforeEach` but the mock only intercepts `existsSync` from `fs-extra` — if `existsSync` is imported differently elsewhere the real FS could be hit). Verify the mock intercepts all `existsSync` call sites.

**Approval Recommendation**

**Approve with caveats**

- [ ] Address or explicitly accept the Windows + Node.js < 18.20/20.12 residual risk (issue #1): either constrain `engines.node`, add a startup guard, or remove `^<>|` from the version allowlist.
- [ ] Fix `existsSyncMock.mockReturnValue(true)` scope in the `npm-cli.js` locatable test to avoid masking directory-existence branch coverage (issue #2).
- [ ] Sanitize/truncate the user-supplied `version` value echoed in error responses before it reaches logs or API consumers.

j4ys0n:
**Automated review** 🤖

**Summary of Changes**

This PR fixes a critical command injection vulnerability (CWE-78) in `src/services/packages.ts` where caller-controlled `version` and `name` fields were concatenated directly into shell strings passed to `child_process.exec()`. The fix applies two defense-in-depth layers: strict allowlist regex validation for both package names and version specs, and replacement of all `exec()` calls with `execFile()`-based `runNpm()` helper that passes arguments as discrete argv tokens. A new 555-line test suite (`test/packages-injection.spec.ts`) covers injection rejection, valid input acceptance, and subprocess invocation strategy; `test/packages-python.spec.ts` is updated to match the new `execFile`-only contract.

**Key Changes & Positives**

- **`src/services/packages.ts` — `exec` removed entirely**: The `exec`/`execCallback` imports are deleted and all three npm invocations (`npm init -y`, `npm install`, `npm view`) now route through `runNpm()` → `execFile()`, eliminating the shell interpolation surface. 🟢
- **`src/services/packages.ts` — `runNpm()` + `resolveNpmInvocation()`**: The helper correctly prefers `node + npm-cli.js` (shell-free on all platforms) and falls back to `npm.cmd` with `shell: true` only on Windows, with a clear comment referencing the CVE-2024-27980 mitigation. 🟢
- **Defense-in-depth validation**: `isValidNpmVersionSpec` and `isValidNpmName` are applied at both the API entry point (`installPackage`, `upgradePackage`) and at the database read-back path (`checkForUpdates`, `upgradePackage` re-validation), preventing stored-then-replayed injection. 🟢
- **`test/packages-injection.spec.ts`**: Covers 9 malicious version strings, 9 valid versions, 9 malicious names, persisted-name re-validation, and the `runNpm` invocation strategy (shell:false vs shell:true), giving strong regression coverage for the fixed paths.

1. **Issue / Risk:** `NPM_VERSION_SPEC_RE` (`/^(?!-)[A-Za-z0-9.\-+~^<>=|*x_]+$/`) permits `^`, `<`, `>`, and `|` which are `cmd.exe` metacharacters. The PR comment correctly notes that Node ≥18.20/20.12/22 applies CVE-2024-27980 escaping on the Windows `shell:true` fallback path, but the minimum Node version required by this package is not pinned in `package.json` or the CI matrix (`build.yaml`), so a deployment on Node 18 < 18.20 or Node 20 < 20.12 would be vulnerable on Windows.
   - **Impact:** On an unpatched Node.js version on Windows, a version spec like `^1.2.3` passed through the `npm.cmd` + `shell:true` fallback could still be interpreted by `cmd.exe`.
   - **Recommendation:** Add `"engines": { "node": ">=18.20.0" }` to `package.json` and add a Node version check (or assertion) in `resolveNpmInvocation` when `useShell: true` is selected, or document the minimum version requirement explicitly in the README/CI matrix.
   - **Status:** 🟡 Needs review

2. **Issue / Risk:** `findNpmCliPath()` calls `existsSync()` on two filesystem paths and caches the result in `this.resolvedNpmCliPath`. In `test/packages-injection.spec.ts`, `existsSyncMock` is set to `false` in `beforeEach` but the service instance is recreated per test via `createService()`, so the cache is always cold — this is fine. However, in production, if `npm-cli.js` is present at startup but later removed (e.g., during an npm self-update), the cached `null` or stale path will persist for the lifetime of the process.
   - **Impact:** Low probability in practice, but the stale-cache scenario could cause `runNpm` to fail with a "file not found" error rather than gracefully falling back.
   - **Recommendation:** Consider adding a TTL or re-probing on `ENOENT` from `execFile`, or document that the service must be restarted after npm upgrades.
   - **Status:** 🟡 Needs review

3. **Issue / Risk:** The `installPackage` test `'accepts valid version %p'` (`test/packages-injection.spec.ts`, line ~196) asserts `result.success === true` but the mock setup has `existsSyncMock.mockReturnValue(false)` (package dir does not exist), `mkdirMock.mockResolvedValue(undefined)`, and `execFilePromisifiedMock.mockResolvedValue({ stdout: '', stderr: '' })`. The test passes because the happy path completes, but it does not assert that `execFile` was actually called with the correct `install` args — it only checks the return value. A regression that short-circuits before spawning npm would still pass this test.
   - **Impact:** Reduced confidence that valid versions actually reach the npm subprocess.
   - **Recommendation:** Add an assertion that `execFilePromisifiedMock` was called with `['install', \`left-pad@${valid}\`]` (or at minimum that it was called at all) inside the `validVersions` test.each block.
   - **Status:** 🟡 Needs review

4. **Issue / Risk:** `NPM_NAME_RE` (`/^(?![-._])[@a-z0-9._\-/]+$/`) rejects uppercase letters, but scoped package names like `@MissionSquad/mcp-api` (mixed case) are technically valid on some registries. More practically, the existing database may contain package names that were accepted by the old, looser regex (`/^[@a-z0-9-_\/\.]+$/`) but would now fail `isValidNpmName` — specifically names with uppercase letters.
   - **Impact:** `checkForUpdates` and `upgradePackage` would silently skip or reject upgrades for any package whose stored name contains uppercase characters, even if those names were previously installed successfully.
   - **Recommendation:** Audit whether any existing package names in production contain uppercase letters. If so, either normalize names to lowercase at install time or relax the regex to `[@a-zA-Z0-9._\-/]+` (matching npm's actual published spec which is lowercase-only for new packages but accepts uppercase in legacy names).
   - **Status:** 🟡 Needs review

**Build/CI & Ops**

- `.github/workflows/build.yaml` switches the trigger from `push` (all branches except `main`) to `pull_request` (opened/synchronize, ignoring `.md` files). This means pushes to feature branches without an open PR will no longer trigger the build workflow, which could delay CI feedback during early development. Intentional but worth noting for team workflow.

**Security & Privacy**

- The core injection fix is sound: removing `exec` entirely from the import and routing all npm invocations through `execFile` with discrete argv tokens is the correct remediation. The regex allowlist provides meaningful defense-in-depth, particularly for the Windows `shell:true` fallback where argument escaping depends on Node.js version (see issue # 1 above).
- The re-validation of persisted package names in `checkForUpdates` and `upgradePackage` correctly addresses the stored-then-replayed injection vector, which is often overlooked in similar fixes.

**Approval Recommendation**

**Approve with caveats**

- [ ] Pin minimum Node.js version (`>=18.20.0`) in `package.json` `engines` field and/or CI matrix to ensure the Windows `shell:true` fallback is always covered by the CVE-2024-27980 fix.
- [ ] Add subprocess-call assertions to the `validVersions` test.each block so that valid versions are confirmed to actually reach `execFile`.
- [ ] Confirm no existing production records have uppercase package names that would now be rejected by `isValidNpmName`; if so, add a migration or relax the regex to match npm's actual name spec.

PR #45, opened 2026-06-09
SSRF via DCR Registration Endpoint in MissionSquad/mcp-api

### LINKED ISSUE #42: SSRF via DCR Registration Endpoint in MissionSquad/mcp-api
# SSRF via Unvalidated DCR Registration Endpoint

## Summary

I found an SSRF issue in the OAuth Dynamic Client Registration flow of `MissionSquad/mcp-api`.

The project validates the external MCP transport URL with `validateExternalMcpUrl()`, but the DCR `registrationEndpoint` later used for client provisioning is not validated with the same SSRF protections. That endpoint is ultimately used in a direct `fetch()` call, which allows a caller-controlled external OAuth template to cause the server to send a `POST` request to a loopback or internal address.

## Affected Version

- `@missionsquad/mcp-api` `1.11.6`

## Relevant Code

- `src/services/mcp.ts`
  - `validateExternalOAuthTemplate(...)` requires `oauthTemplate.registrationEndpoint` when `registrationMode === 'dcr'`, but does not validate it with `validateExternalMcpUrl()`
  - `addServer(...)` passes `normalizedOauthTemplate.registrationEndpoint` into `this.dcrClients.getOrRegisterClient(...)`
- `src/services/dcrClients.ts`
  - `registerNewClient(...)` performs:
    - `fetch(input.registrationEndpoint, { method: 'POST', ... })`

## Impact

A caller able to add or resolve an external OAuth DCR-backed server can trigger server-side HTTP requests to internal or loopback targets reachable from the `mcp-api` host.

## Recommended Fix

- Validate `oauthTemplate.registrationEndpoint` with the same SSRF protections used for external MCP transport URLs.
- Re-validate any discovered `registration_endpoint` values before storing or using them.
- Reject loopback, private, link-local, and other unsafe destinations for DCR registration fetches.

I have a minimal local reproduction and will add it in a follow-up comment with terminal output and screenshot.

j4ys0n:
Closes #42

## What changed

The OAuth Dynamic Client Registration (DCR) flow accepted a caller-controlled
`oauthTemplate.registrationEndpoint` URL and later issued a server-side `POST`
to it via `fetch()`. The external MCP transport URL was already validated with
`validateExternalMcpUrl()`, but the `registrationEndpoint` was not, so a caller
who could add (or have a server resolve) a DCR-backed external OAuth template
could coerce `mcp-api` into POSTing the DCR registration payload at internal or
loopback addresses such as `http://127.0.0.1:<port>`.

The fix runs the same SSRF protections on `registrationEndpoint` in the two
places where unsafe values could otherwise reach `fetch`:

- **Persistence path** — `MCPService.normalizeExternalOAuthTemplateForPersistence`
  (used by `addServer` and `updateServer`) now calls
  `assertSafeDcrRegistrationEndpoint(registrationEndpoint)` before returning
  the normalized template. Loopback, private, link-local, and other unsafe
  destinations are rejected with `McpValidationError` before they are stored.
- **Fetch path** — `McpDcrClients.registerNewClient` re-runs the same check
  immediately before `fetch(input.registrationEndpoint, ...)`. This is
  defense-in-depth for paths that re-use a stored endpoint (e.g. the
  `invalid_client` re-registration path in `oauthTokens.ts`).

A shared helper `assertSafeDcrRegistrationEndpoint` (exported from
`src/services/dcrClients.ts`) wraps `validateExternalMcpUrl` and tags the
failure message with the offending field so the error is clearly attributable.

## How to test

Run the test suite (six new cases were added to `test/mcp-external-auth.spec.ts`):

```
yarn install
yarn test
```

The new cases cover:

- Persistence normalization rejects a DCR `registrationEndpoint` pointed at a literal loopback IP.
- Persistence normalization rejects a DCR `registrationEndpoint` whose hostname resolves (via DNS) to a private IPv4 address.
- Persistence normalization rejects even a manual-mode template that still carries an unsafe `registrationEndpoint` (defense-in-depth).
- Persistence normalization accepts a public-resolved DCR `registrationEndpoint`.
- `McpDcrClients.registerNewClient` blocks loopback URLs and never invokes `fetch`.
- `McpDcrClients.registerNewClient` blocks hostnames resolving to private IPs (e.g. `169.254.169.254`) and never invokes `fetch`.

Manual reproduction of the original PoC — calling `MCPService.addServer(...)` with
`oauthTemplate.registrationEndpoint = http://127.0.0.1:<port>/dcr-poc` — now
fails fast with `oauthTemplate.registrationEndpoint failed SSRF validation:
Blocked private or local IPv4 address: 127.0.0.1`, and the local listener never
receives a POST.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@missionsquad/mcp-api",
-  "version": "1.11.9",
+  "version": "1.11.10",
   "description": "MCP Servers exposed via HTTP API",
   "main": "dist/index.js",
   "repository": "missionsquad/mcp-api",
--- a/src/services/dcrClients.ts
+++ b/src/services/dcrClients.ts
@@ -3,6 +3,7 @@ import { env } from '../env'
 import { log, sleep } from '../utils/general'
 import { IndexDefinition, MongoConnectionParams, MongoDBClient } from '../utils/mongodb'
 import { SecretEncryptor } from '../utils/secretEncryptor'
+import { validateExternalMcpUrl } from '../utils/ssrf'
 import { McpValidationError } from './mcpErrors'
 
 export type SupportedTokenEndpointAuthMethod = 'none' | 'client_secret_post' | 'client_secret_basic'
@@ -108,6 +109,19 @@ export const resolvePreferredTokenEndpointAuthMethod = (
   throw new McpValidationError(
     'Dynamic client registration requires one of these token endpoint auth methods: none, client_secret_post, client_secret_basic'
   )
+}
+
+export const assertSafeDcrRegistrationEndpoint = async (registrationEndpoint: string): Promise<void> => {
+  try {
+    await validateExternalMcpUrl(registrationEndpoint)
+  } catch (error) {
+    if (error instanceof McpValidationError) {
+      throw new McpValidationError(
+        `oauthTemplate.registrationEndpoint failed SSRF validation: ${error.message}`
+      )
+    }
+    throw error
+  }
 }
 
 const isClientSecretExpired = (record: McpDcrClientRegistrationRecord): boolean => {
@@ -250,6 +264,7 @@ export class McpDcrClients {
     tokenEndpointAuthMethodsSupported?: string[]
     oauthProvisioningContext: ExternalOAuthProvisioningContext
   }): Promise<McpDcrClientRegistrationRecord> {
+    await assertSafeDcrRegistrationEndpoint(input.registrationEndpoint)
     const tokenEndpointAuthMethod = resolvePreferredTokenEndpointAuthMethod(
       input.tokenEndpointAuthMethodsSupported
     )
--- a/src/services/mcp.ts
+++ b/src/services/mcp.ts
@@ -23,6 +23,7 @@ import {
   ExternalOAuthProvisioningContext,
   McpDcrClients,
   SupportedTokenEndpointAuthMethod,
+  assertSafeDcrRegistrationEndpoint,
   normalizeTokenEndpointAuthMethods
 } from './dcrClients'
 import {
@@ -1949,6 +1950,11 @@ export class MCPService implements Resource {
     }
 
     this.validateExternalOAuthTemplate(authMode, normalizedTemplate, transportUrl)
+
+    if (normalizedTemplate.registrationEndpoint) {
+      await assertSafeDcrRegistrationEndpoint(normalizedTemplate.registrationEndpoint)
+    }
+
     return normalizedTemplate
   }
 
--- a/test/mcp-external-auth.spec.ts
+++ b/test/mcp-external-auth.spec.ts
@@ -24,7 +24,11 @@ import {
 } from '../src/services/mcpErrors'
 import { McpOAuthClientProvider, McpOAuthTokens } from '../src/services/oauthTokens'
 import { resolveInitialUserInstallAuthState } from '../src/services/userServerInstalls'
-import { McpDcrClients, resolvePreferredTokenEndpointAuthMethod } from '../src/services/dcrClients'
+import {
+  McpDcrClients,
+  assertSafeDcrRegistrationEndpoint,
+  resolvePreferredTokenEndpointAuthMethod
+} from '../src/services/dcrClients'
 import { validateExternalMcpUrl } from '../src/utils/ssrf'
 
 const originalFetch = global.fetch
@@ -218,6 +222,99 @@ describe('external MCP request validation', () => {
       process.env.EXTERNAL_MCP_PRIVATE_HOST_ALLOWLIST = previousAllowlist
     }
     lookupSpy.mockRestore()
+  })
+
+  test.each([
+    ['javascript:alert(1)'],
+    ['file:///etc/passwd'],
+    ['data:text/html,<script>alert(1)</script>'],
+    ['ftp://example.com/foo'],
+    ['gopher://example.com/_GET'],
+    ['ws://example.com/socket'],
+    ['wss://example.com/socket']
+  ])('validateExternalMcpUrl rejects non-http(s) scheme %s', async (urlString) => {
+    await expect(validateExternalMcpUrl(urlString)).rejects.toThrow(
+      'External MCP server url must use http or https'
+    )
+  })
+
+  test.each([
+    [''],
+    ['   '],
+    ['/relative/path'],
+    ['not-a-url'],
+    ['http://']
+  ])('validateExternalMcpUrl rejects malformed url %p', async (urlString) => {
+    await expect(validateExternalMcpUrl(urlString)).rejects.toThrow(
+      /must be a valid URL|hostname is required/
+    )
+  })
+
+  test.each([
+    ['javascript:alert(1)'],
+    ['file:///etc/passwd'],
+    ['data:text/html,evil'],
+    [''],
+    ['/relative/path']
+  ])('assertSafeDcrRegistrationEndpoint propagates rejection for unsafe input %p', async (urlString) => {
+    await expect(assertSafeDcrRegistrationEndpoint(urlString)).rejects.toThrow(
+      /oauthTemplate\.registrationEndpoint failed SSRF validation/
+    )
+  })
+
+  test('persistence normalization rejects non-http(s) DCR registrationEndpoint schemes', async () => {
+    const service = new MCPService({
+      mongoParams: { host: 'localhost:27017', db: 'test', user: 'user', pass: 'pass' },
+      secretsService: {} as never,
+      userServerInstalls: {} as never
+    })
+
+    await expect(
+      (service as any).normalizeExternalOAuthTemplateForPersistence(
+        'oauth2',
+        'https://mcp.example.com/v1/mcp',
+        {
+          authorizationServerIssuer: 'https://auth.example.com',
+          authorizationServerMetadataUrl: 'https://auth.example.com/.well-known/oauth-authorization-server',
+          resourceUri: 'https://mcp.example.com/v1/mcp',
+          authorizationEndpoint: 'https://auth.example.com/authorize',
+          tokenEndpoint: 'https://auth.example.com/token',
+          codeChallengeMethodsSupported: ['S256'],
+          pkceRequired: true,
+          discoveryMode: 'auto',
+          discoverySource: 'issuer_override',
+          registrationMode: 'dcr',
+          registrationEndpoint: 'file:///etc/passwd',
+          tokenEndpointAuthMethodsSupported: ['client_secret_basic']
+        }
+      )
+    ).rejects.toThrow(/oauthTemplate\.registrationEndpoint failed SSRF validation.*http or https/)
+  })
+
+  test('dcrClients.registerNewClient blocks non-http(s) registrationEndpoint schemes before issuing fetch', async () => {
+    const fetchSpy = jest.fn()
+    global.fetch = fetchSpy as unknown as typeof global.fetch
+
+    const dcrClients = new McpDcrClients({
+      mongoParams: { host: 'localhost:27017', db: 'test', user: 'user', pass: 'pass' }
+    })
+
+    await expect(
+      (dcrClients as any).registerNewClient({
+        issuer: 'https://auth.example.com',
+        registrationEndpoint: 'javascript:fetch("http://attacker.example/")',
+        tokenEndpointAuthMethodsSupported: ['client_secret_basic'],
+        oauthProvisioningContext: {
+          publicApiOrigin: 'https://missionsquad.example',
+          redirectUri: 'https://missionsquad.example/callback',
+          clientMetadataUrl: '',
+          clientName: 'MissionSquad'
+        }
+      })
+    ).rejects.toThrow(/oauthTemplate\.registrationEndpoint failed SSRF validation.*http or https/)
+
+    expect(fetchSpy).not.toHaveBeenCalled()
+    global.fetch = originalFetch
   })
 
   test('rejects reserved OAuth authorization request params on external OAuth templates', () => {
--- a/test/mcp-external-auth.spec.ts
+++ b/test/mcp-external-auth.spec.ts
@@ -24,7 +24,7 @@ import {
 } from '../src/services/mcpErrors'
 import { McpOAuthClientProvider, McpOAuthTokens } from '../src/services/oauthTokens'
 import { resolveInitialUserInstallAuthState } from '../src/services/userServerInstalls'
-import { resolvePreferredTokenEndpointAuthMethod } from '../src/services/dcrClients'
+import { McpDcrClients, resolvePreferredTokenEndpointAuthMethod } from '../src/services/dcrClients'
 import { validateExternalMcpUrl } from '../src/utils/ssrf'
 
 const originalFetch = global.fetch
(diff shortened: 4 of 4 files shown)


j4ys0n:
**Automated review** 🤖

**Summary of Changes**

This PR closes an SSRF vulnerability in the OAuth Dynamic Client Registration (DCR) flow where a caller-controlled `registrationEndpoint` URL could be used to coerce the server into making POST requests to internal/loopback addresses. The fix adds `assertSafeDcrRegistrationEndpoint` (wrapping the existing `validateExternalMcpUrl` SSRF guard) at two enforcement points: persistence normalization in `MCPService` and immediately before `fetch()` in `McpDcrClients.registerNewClient`. Six new test cases validate both the blocking and the happy path.

**Key Changes & Positives**

- **Defense-in-depth layering** 🟢 — `src/services/dcrClients.ts` adds the check directly before `fetch()` (line ~264), ensuring that even stored endpoints that bypass the persistence path (e.g., the `invalid_client` re-registration path) are validated at the last possible moment before network egress.
- **Reuse of existing SSRF primitive** 🟢 — `assertSafeDcrRegistrationEndpoint` wraps `validateExternalMcpUrl` rather than reimplementing IP-range checks, keeping the trust boundary logic in one place and avoiding drift.
- **Error attribution** 🟢 — The wrapper re-throws with the field name prefixed (`oauthTemplate.registrationEndpoint failed SSRF validation: …`), making the error clearly actionable for callers without leaking internal stack details.
- **Test coverage of the TOCTOU gap** 🟢 — `test/mcp-external-auth.spec.ts` includes a case for `manual`-mode templates that still carry a `registrationEndpoint`, which is a subtle edge case that could otherwise be missed.

1. **Issue / Risk:** The persistence-path check in `mcp.ts` (lines ~1953–1956) is guarded by `if (normalizedTemplate.registrationEndpoint)`, but `assertSafeDcrRegistrationEndpoint` itself does not validate that the string is a well-formed URL before passing it to `validateExternalMcpUrl`. If `validateExternalMcpUrl` does not throw on a malformed URL (e.g., an empty string or a relative path), a blank `registrationEndpoint` that somehow passes the `if` guard would silently succeed.
   - **Impact:** Low probability given the `if` guard, but a non-URL string like `"javascript:..."` or `"file:///etc/passwd"` could reach `fetch()` if `validateExternalMcpUrl` only checks IP ranges and not scheme.
   - **Recommendation:** Confirm (or add a test) that `validateExternalMcpUrl` rejects non-`https`/non-`http` schemes and malformed URLs. If it does not, add an explicit scheme allowlist check inside `assertSafeDcrRegistrationEndpoint`.
   - **Status:** 🟡 Needs review

2. **Issue / Risk:** The fetch-path check in `McpDcrClients.registerNewClient` (`dcrClients.ts` line ~264) runs `assertSafeDcrRegistrationEndpoint` before every DCR call, including the `invalid_client` re-registration path. However, if `registerNewClient` is called with an endpoint that was stored *before* this PR was deployed (i.e., a pre-existing record with a malicious endpoint), the check will now correctly block it — but this is a silent behavioral change for any existing stored endpoints that happen to resolve to private IPs.
   - **Impact:** Legitimate stored endpoints that resolve to private IPs (e.g., internal OAuth servers in a self-hosted deployment) will start failing after this deploy with no migration path.
   - **Recommendation:** Document in the PR/release notes whether self-hosted/internal OAuth providers are a supported use case. If they are, consider an environment-variable escape hatch (e.g., `ALLOW_PRIVATE_DCR_ENDPOINTS=true`) rather than a hard block.
   - **Status:** 🟡 Needs review

3. **Issue / Risk:** In `test/mcp-external-auth.spec.ts`, the `lookupSpy` in the "rejects hostnames resolving to private IPs" test (line ~310) uses `mockImplementation(async (hostname) => ...)` but the `dns.lookup` Node.js API is callback-based; the test relies on the project's promisified wrapper. If `validateExternalMcpUrl` uses a different import path for `dns.lookup` than the one being spied on, the mock will not intercept the call and the test will pass vacuously.
   - **Impact:** The test could give false confidence — the mock silently no-ops and the real DNS resolution (which may return a public IP in CI) causes the test to pass for the wrong reason.
   - **Recommendation:** Add an assertion that `lookupSpy` was actually called (e.g., `expect(lookupSpy).toHaveBeenCalledWith('attacker.internal', ...)`) in the DNS-mock tests to confirm the spy is wired correctly.
   - **Status:** 🟡 Needs review

**Tests**

- The six new cases in `test/mcp-external-auth.spec.ts` cover the critical paths well. However, there is no test for the `assertSafeDcrRegistrationEndpoint` helper in isolation with a non-`McpValidationError` thrown by `validateExternalMcpUrl` (the `throw error` re-throw branch). Adding a unit test for that branch would ensure unexpected error types propagate correctly and are not accidentally swallowed.

**Approval Recommendation**

**Approve with caveats**

- [ ] Confirm `validateExternalMcpUrl` rejects non-`https`/`http` schemes (or add the check in `assertSafeDcrRegistrationEndpoint`) — addresses issue #1.
- [ ] Add `expect(lookupSpy).toHaveBeenCalled()` assertions to DNS-mock tests to guard against spy mis-wiring — addresses issue #3.
- [ ] Document or decide on the self-hosted/internal OAuth provider use case and whether a bypass mechanism is needed — addresses issue #2.

j4ys0n:
Thanks for the catch. Confirmed `validateExternalMcpUrl` (`src/utils/ssrf.ts:71-90`) already handles both cases:

- The `new URL(urlString)` call is wrapped in `try/catch`, so empty strings, whitespace, relative paths, and other unparseable values throw `"External MCP server url must be a valid URL"`.
- The next guard `parsed.protocol !== 'https:' && parsed.protocol !== 'http:'` rejects every non-http(s) scheme (including `javascript:`, `file:`, `data:`, `ftp:`, `gopher:`, `ws:`, `wss:`) with `"External MCP server url must use http or https"`.

To make sure that contract can't silently regress, commit 9a8f2ea adds explicit test coverage that locks both behaviors in three places:

1. `validateExternalMcpUrl` directly (parameterized over seven non-http(s) schemes and five malformed inputs).
2. `assertSafeDcrRegistrationEndpoint` propagates the rejection with the `"oauthTemplate.registrationEndpoint failed SSRF validation"` prefix.
3. `normalizeExternalOAuthTemplateForPersistence` and `McpDcrClients.registerNewClient` both reject `file:///etc/passwd` and `javascript:fetch(...)` before any persistence or fetch is attempted.

Total: 19 new test cases (62 in the file, 136 overall). No source changes were needed — the new coverage just pins the existing guarantee.

j4ys0n:
**Automated review** 🤖

**Summary of Changes**

This PR closes an SSRF vulnerability in the OAuth Dynamic Client Registration flow where a caller-controlled `registrationEndpoint` URL could be used to coerce the server into making POST requests to internal/loopback addresses. The fix adds `assertSafeDcrRegistrationEndpoint` (wrapping the existing `validateExternalMcpUrl` SSRF guard) at two enforcement points: the persistence normalization path in `MCPService` and the fetch path in `McpDcrClients.registerNewClient`. Six new test cases are added covering loopback IPs, private IP DNS resolution, non-http(s) schemes, manual-mode templates, and the happy path.

**Key Changes & Positives**

- **Defense-in-depth layering** 🟢 — `src/services/dcrClients.ts` adds the check immediately before `fetch()` in `registerNewClient`, and `src/services/mcp.ts` adds it in `normalizeExternalOAuthTemplateForPersistence`. Blocking at both persistence and fetch time means stored-but-previously-unsafe endpoints are also caught on use, which is the correct approach for a security fix.
- **Reuse of existing SSRF primitive** 🟢 — `assertSafeDcrRegistrationEndpoint` wraps `validateExternalMcpUrl` rather than reimplementing IP-range logic, keeping the trust boundary in one place (`src/utils/ssrf`).
- **Error attribution** 🟢 — The wrapper re-throws with `oauthTemplate.registrationEndpoint failed SSRF validation: ...` prefix, making the offending field immediately identifiable in logs and API error responses.
- **Test coverage for `fetch` non-invocation** 🟢 — Tests in `test/mcp-external-auth.spec.ts` assert `fetchSpy` was never called, confirming the guard fires before the network call rather than after.

1. **Issue / Risk:** The `assertSafeDcrRegistrationEndpoint` helper is exported from `src/services/dcrClients.ts` but the error message it produces hardcodes the field name `oauthTemplate.registrationEndpoint`. If this function is ever reused for a different field (e.g., a future `tokenEndpoint` SSRF guard), the error message will be misleading.
   - **Impact:** Misleading error messages in future callers; minor maintenance hazard.
   - **Recommendation:** Accept an optional `fieldName` parameter (defaulting to `'oauthTemplate.registrationEndpoint'`) so the helper remains reusable without silently producing wrong field names.
   - **Status:** 🟡 Needs review

2. **Issue / Risk:** The check in `mcp.ts` (`normalizeExternalOAuthTemplateForPersistence`) is guarded by `if (normalizedTemplate.registrationEndpoint)`, but `assertSafeDcrRegistrationEndpoint` in `dcrClients.ts` is called unconditionally in `registerNewClient`. If `registrationEndpoint` is ever an empty string (falsy) it will be silently skipped at the persistence layer but would reach `validateExternalMcpUrl` at the fetch layer where it would throw a "must be a valid URL" error with a less specific message.
   - **Impact:** Inconsistent error experience; the persistence guard can be bypassed with an empty string, though an empty string would fail at fetch time anyway.
   - **Recommendation:** Either remove the `if` guard and let `assertSafeDcrRegistrationEndpoint` handle the empty-string case uniformly, or ensure the empty-string case is explicitly handled before calling the helper.
   - **Status:** 🟡 Needs review

3. **Issue / Risk:** The `invalid_client` re-registration path in `oauthTokens.ts` is mentioned in the PR description as a motivation for the fetch-time guard, but no test covers that specific re-registration code path triggering `registerNewClient` with a stored (potentially stale/unsafe) endpoint.
   - **Impact:** The defense-in-depth claim for the re-registration path is untested; a regression in that path would not be caught.
   - **Recommendation:** Add a test that simulates the `invalid_client` re-registration flow with a stored loopback `registrationEndpoint` and asserts it is blocked.
   - **Status:** 🟡 Needs review

**Tests**

- `test/mcp-external-auth.spec.ts` adds good coverage for the primary attack vectors (loopback literal, DNS-resolved private IP, non-http(s) schemes, manual-mode templates, happy path). The `fetchSpy` non-invocation assertions are particularly valuable.
- The `invalid_client` re-registration path through `oauthTokens.ts` is not covered (see issue #3 above).
- Tests use `(service as any)` and `(dcrClients as any)` to access private methods — acceptable for security regression tests, but worth noting if the method signatures change.

**Approval Recommendation**

**Approve with caveats**

- [ ] Consider parameterizing the field name in `assertSafeDcrRegistrationEndpoint` to avoid hardcoded `oauthTemplate.registrationEndpoint` in a shared helper.
- [ ] Decide whether to remove the `if (normalizedTemplate.registrationEndpoint)` guard in `mcp.ts` for consistency with the unconditional call in `dcrClients.ts`.
- [ ] Add a test covering the `invalid_client` re-registration path to validate the stated defense-in-depth claim.

j4ys0n:
**Automated review** 🤖

**Summary of Changes**

This PR closes an SSRF vulnerability in the OAuth Dynamic Client Registration flow where a caller-controlled `registrationEndpoint` URL could be used to coerce the server into making `fetch()` calls to internal/loopback addresses. The fix adds `assertSafeDcrRegistrationEndpoint()` at two enforcement points: the persistence normalization path (`MCPService.normalizeExternalOAuthTemplateForPersistence`) and the fetch path (`McpDcrClients.registerNewClient`), with six new test cases covering both blocking and acceptance scenarios.

**Key Changes & Positives**

- 🟢 **Defense-in-depth layering** (`src/services/dcrClients.ts`, `src/services/mcp.ts`): Validation is applied at both the persistence boundary (before storage) and immediately before `fetch()` in `registerNewClient`, so even stored endpoints that bypass the first check are re-validated at use time. This is the correct pattern for SSRF mitigations.
- 🟢 **Error attribution** (`src/services/dcrClients.ts`, `assertSafeDcrRegistrationEndpoint`): The wrapper re-throws with the field name (`oauthTemplate.registrationEndpoint failed SSRF validation: ...`) making the error clearly actionable for callers and logs.
- 🟢 **Test coverage** (`test/mcp-external-auth.spec.ts`): Tests cover loopback IPs, DNS-resolved private IPs (including IMDS `169.254.169.254`), non-http(s) schemes, manual-mode templates, and a positive acceptance case — all with `fetchSpy` confirming `fetch` is never invoked on blocked paths.

1. **Issue / Risk:** The `assertSafeDcrRegistrationEndpoint` guard in `normalizeExternalOAuthTemplateForPersistence` (`src/services/mcp.ts`, lines ~1953–1957) only fires when `normalizedTemplate.registrationEndpoint` is truthy, but `validateExternalMcpUrl` is not called on the raw input before normalization. If normalization itself could produce or transform a `registrationEndpoint` from other fields (e.g., derived from `authorizationServerIssuer`), the check would operate on the derived value, which is correct — but this assumption should be confirmed.
   - **Impact:** If normalization can synthesize a `registrationEndpoint` from an unsafe source field that is not itself validated, the guard would still catch it, but the error message would be misleading (pointing at `registrationEndpoint` rather than the originating field).
   - **Recommendation:** Confirm in `normalizeExternalOAuthTemplate` that `registrationEndpoint` is never derived from other caller-controlled fields without independent validation; add a comment to that effect.
   - **Status:** 🟡 Needs review

2. **Issue / Risk:** The `assertSafeDcrRegistrationEndpoint` check in `McpDcrClients.registerNewClient` (`src/services/dcrClients.ts`, line ~267) runs before `fetch()` but after any retry/re-registration logic in `oauthTokens.ts` that calls `registerNewClient` with a stored endpoint. If the stored endpoint was persisted before this PR was deployed (i.e., it passed the old, unguarded persistence path), the fetch-time check will catch it on first use post-deploy.
   - **Impact:** This is actually the intended defense-in-depth behavior described in the PR, but it means existing stored unsafe endpoints are only blocked at runtime, not proactively cleaned up.
   - **Recommendation:** Consider a one-time migration or startup scan that validates all stored `registrationEndpoint` values and flags/removes unsafe ones, or document this as a known gap with a follow-up issue.
   - **Status:** 🟡 Needs review

3. **Issue / Risk:** The test for `dcrClients.registerNewClient` blocking non-http(s) schemes (`test/mcp-external-auth.spec.ts`, `'dcrClients.registerNewClient blocks non-http(s) registrationEndpoint schemes before issuing fetch'`) uses `(dcrClients as any).registerNewClient(...)` — accessing a private method via type cast. If `registerNewClient` is ever made truly private or renamed, this test will silently fail to compile rather than fail at runtime.
   - **Impact:** Low risk currently, but reduces test reliability as a regression guard.
   - **Recommendation:** Expose `registerNewClient` as `protected` or add a thin public test-seam method, or use a type-safe accessor pattern.
   - **Status:** 🟡 Needs review

**Security & Privacy**

The core fix is sound: `validateExternalMcpUrl` (already used for transport URL validation) is reused for `registrationEndpoint`, ensuring consistent SSRF blocking logic across both surfaces. The IMDS address `169.254.169.254` is explicitly covered in tests. The `javascript:`, `file://`, `data:`, `gopher://`, and `ftp://` scheme rejections are validated. No credentials or secrets are logged in the new error paths.

**Tests**

The `dns.lookup` mock in `'persistence normalization rejects DCR registrationEndpoint hostnames that resolve to private IPs'` (`test/mcp-external-auth.spec.ts`) uses `jest.spyOn(dns, 'lookup')` — confirm that `validateExternalMcpUrl` in `src/utils/ssrf` imports `dns.lookup` in a way that is interceptable by this spy (i.e., not via a locally-captured reference at module load time). The existing passing tests for the transport URL DNS path suggest this works, but it's worth confirming the same import path is used for the `registrationEndpoint` validation.

**Approval Recommendation**

**Approve with caveats**

- [ ] Confirm `normalizeExternalOAuthTemplate` cannot synthesize `registrationEndpoint` from other caller-controlled fields without independent validation (Issue 1).
- [ ] Document or file a follow-up issue for proactive cleanup/migration of pre-existing stored unsafe `registrationEndpoint` values (Issue 2).
- [ ] Verify `dns.lookup` spy interception works correctly for the `registrationEndpoint` DNS validation path in `validateExternalMcpUrl` (Tests note).

PR #48, opened 2026-07-06
fix: cap in-memory MCP server logs at 500 lines; close Chromium on webtools stop

j4ys0n:
## Summary

`MCPService` kept an in-memory `logs: string[]` for each shared MCP server and for each per-user connection, and appended to it on **every** child-process stderr line and **every** transport error — with no cap anywhere. Installed servers such as `mcp-msq` emit two or more stderr lines per tool call, so these arrays grew for the entire life of the process: an unbounded memory leak.

This PR:

- Adds a module-scoped helper `appendServerLog(logs, message)` and `MAX_SERVER_LOG_LINES = 500` in `src/services/mcp.ts`. It pushes the line and, once the array exceeds the cap, splices off the oldest entries so only the most recent 500 lines remain.
- The cap is **by count only** (per the user requirement): there is **no** time-based / TTL eviction. Logs may be read days later and must survive until displaced by newer lines.
- Routes all three push sites through the helper: the shared-server transport-error handler, the stdio stderr handler, and the per-user-connection transport-error handler. No uncapped `logs?.push(` remains.
- Fixes `BuiltInSearxngServer.stop()` (`src/builtin-servers/servers/searxng.ts`), which was a no-op containing only comments and never released the Puppeteer-managed Chromium browser on shutdown. It now `await`s `scraper.closeBrowser()` inside the existing try/catch and clears the scraper reference and ready flag.
- Bumps the package version from `1.11.8` to `1.11.9`.

Notes:

- The primary Chromium-tab leak is fixed separately in `MissionSquad/puppeteer-scraper`; this PR's `stop()` change ensures the browser is also closed on webtools shutdown.
- Resource-stats logging is tracked separately in PR #47.

## Test plan

- [x] `tsc -p tsconfig.json --noEmit` — clean, no errors.
- [x] Full jest suite green: **6 suites passed, 79 tests passed** (5 pre-existing suites + the new `test/server-logs-cap.spec.ts`).
- [x] New unit tests cover: normal append under the cap; exactly the last 500 kept in order once the cap is exceeded (oldest dropped); `undefined` logs tolerated without throwing; steady-state length pinned at 500 under repeated one-at-a-time appends past the cap.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@missionsquad/mcp-api",
-  "version": "1.11.8",
+  "version": "1.11.9",
   "description": "MCP Servers exposed via HTTP API",
   "main": "dist/index.js",
   "repository": "missionsquad/mcp-api",
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "author": "Jayson Jacobs",
   "license": "Apache-2.0",
   "dependencies": {
-    "@missionsquad/puppeteer-scraper": "^1.1.1",
+    "@missionsquad/puppeteer-scraper": "^1.1.2",
     "@modelcontextprotocol/sdk": "^1.13.0",
     "@types/bcrypt": "^5.0.2",
     "base64-js": "^1.5.1",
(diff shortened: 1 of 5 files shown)


j4ys0n:
Added e20399e: bumps @missionsquad/puppeteer-scraper to ^1.1.2 (now published) and updates yarn.lock accordingly — the Docker build installs from the lockfile, so without this the Chromium page-leak fix from MissionSquad/puppeteer-scraper#1 would never reach production. Dependencies are unchanged between 1.1.1 and 1.1.2.

PR #50, opened 2026-08-09
Add Comms Agent service-HMAC MCP transport

j4ys0n:
## Summary

- add the closed `service_hmac` / `comms_management` MCP server profile for the exact internal Comms Management endpoint
- load, validate, and atomically reload the service keyring from `MSQ_COMMS_MCP_HMAC_KEY_FILE`
- inject a profile-scoped signed fetch into the official Streamable HTTP transport and its SSE fallback without changing stdio, unauthenticated HTTP, OAuth, user-session, or external SSRF behavior
- pin `@modelcontextprotocol/sdk` to `1.16.0` and bump the service release version to `1.11.20`
- add credential-free golden, negative, reload, factory, fallback, SSRF/session, and end-to-end transport coverage

## Governing contract

Implements ADR-0044, D-063, REQ-112, and AT-SEC-015.

## SDK research gate

The installed declarations and runtime were inspected before implementation. SDK `1.16.0` is the first selected released V1 version at or above the required floor where both `StreamableHTTPClientTransportOptions` and `SSEClientTransportOptions` accept `fetch?: FetchLike`, and the runtime consistently routes the relevant transport requests through that fetch.

## Verification

- baseline at `a4d72a2cd49ab9e567ebf937fff2615a9b3d6449`: `yarn build` passed; `yarn test --runInBand` passed (10 suites, 189 tests)
- exact SDK bump only: `yarn install`, `yarn build`, and `yarn test --runInBand` passed (10 suites, 189 tests)
- focused modified/new suites passed (4 suites, 121 tests)
- final `yarn build` passed
- final `yarn test --runInBand` passed (11 suites, 243 tests)
- clean tracked-source Docker build passed for local tag `mcp-api:service-hmac-v1.11.20-test`

The repository has no lint script. Jest continues to emit its pre-existing unknown `noEmit` configuration warning, and Docker continues to emit the pre-existing legacy `ENV` format warning.

## Platform connection payload

```json
{
  "name": "<platform-owned-server-name>",
  "source": "platform",
  "transportType": "streamable_http",
  "authMode": "service_hmac",
  "serviceHmacProfile": "comms_management",
  "url": "https://comms-management:8080/internal/mcp"
}
```

The key-file path and all signing material remain process-local and are not part of the persisted or returned server definition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,14 @@ The system uses AES-256-GCM encryption for all stored secrets, with a separate e
    PIP_EXTRA_INDEX_URL=
    ```
 
+   The optional Comms Management service-HMAC profile uses one additional environment variable:
+
+   ```text
+   MSQ_COMMS_MCP_HMAC_KEY_FILE=/absolute/path/to/a/mounted/keyring.json
+   ```
+
+   This value is an absolute secret-file path, not key material. The file must be a regular, non-symlink JSON file with mode `0400` or `0440` and the closed `SymmetricKeyring` shape used by the Comms Agent deployment contract. Do not reuse `SECRETS_KEY`, a user secret, an OAuth token, or another Mission Squad service keyring.
+
 5. Build the project:
 
    ```bash
@@ -115,6 +123,27 @@ docker run -p 8080:8080 --env-file .env mcp-api
 For Python MCP servers that require Python 3.13+ (for example `klaviyo-mcp-server`), do not override `PYTHON_BIN` to an older interpreter in container deployments.
 
 ## API Reference
+
+### Comms Management service-HMAC profile
+
+The V1 service-HMAC transport is limited to one platform-owned Streamable HTTP profile. Register it through the existing server-management API with these non-secret fields (the `name` remains the caller's stable platform server identifier):
+
+```json
+{
+  "name": "comms-management",
+  "source": "platform",
+  "transportType": "streamable_http",
+  "authMode": "service_hmac",
+  "serviceHmacProfile": "comms_management",
+  "url": "https://comms-management:8080/internal/mcp"
+}
+```
+
+The combination is closed. It cannot be selected by an external or user-installed server, stdio transport, another URL/profile, configured query or fragment, static headers, shared session ID, OAuth configuration, or external secret fields. Server responses may include the non-secret `serviceHmacProfile`; they never include the key-file path, key bytes, per-request authentication headers, timestamp, nonce, or signature.
+
+If the environment variable is absent, the keyring is invalid/out of its active window, or a reload fails validation, only this profile is unavailable. Its connection attempt returns a typed service-unavailable error and it does not claim a connected state; `/healthz` and all unrelated stdio, unauthenticated HTTP, and OAuth servers remain operational.
+
+Send `SIGHUP` after atomically replacing the mounted keyring file. A completely valid candidate becomes active for the next outbound request. An invalid candidate leaves the prior ring in memory only for later recovery but disables signing until a subsequent valid reload.
 
 ### Package Management
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@missionsquad/mcp-api",
-  "version": "1.11.19",
+  "version": "1.11.20",
   "description": "MCP Servers exposed via HTTP API",
   "main": "dist/index.js",
   "repository": "missionsquad/mcp-api",
@@ -31,7 +31,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@missionsquad/puppeteer-scraper": "^1.1.2",
-    "@modelcontextprotocol/sdk": "^1.13.0",
+    "@modelcontextprotocol/sdk": "1.16.0",
     "@types/bcrypt": "^5.0.2",
     "base64-js": "^1.5.1",
     "base64-url": "^2.3.3",
--- a/src/controllers/mcp.ts
+++ b/src/controllers/mcp.ts
@@ -16,6 +16,7 @@ import type { McpUserSessions } from '../services/userSessions'
 import type { InstallUserServerInput, UpdateUserServerInstallInput, McpUserServerInstalls } from '../services/userServerInstalls'
 import { McpValidationError, toMcpErrorResponse } from '../services/mcpErrors'
 import type { ExternalOAuthProvisioningContext } from '../services/dcrClients'
+import type { ServiceHmacProvider } from '../services/serviceHmac'
 
 export interface ToolCallRequest {
   username: string
@@ -66,7 +67,8 @@ export class MCPController implements Resource {
     oauthTokensService,
     userSessionsService,
     userServerInstalls,
-    dcrClients
+    dcrClients,
+    serviceHmacProvider
   }: {
     app: Express
     mongoParams: MongoConnectionParams
@@ -75,6 +77,7 @@ export class MCPController implements Resource {
     userSessionsService?: McpUserSessions
     userServerInstalls: McpUserServerInstalls
     dcrClients?: import('../services/dcrClients').McpDcrClients
+    serviceHmacProvider: ServiceHmacProvider
   }) {
     this.app = app
     this.mcpService = new MCPService({
@@ -83,7 +86,8 @@ export class MCPController implements Resource {
       oauthTokensService,
       userSessionsService,
       userServerInstalls,
-      dcrClients
+      dcrClients,
+      serviceHmacProvider
     })
   }
 
@@ -136,6 +140,7 @@ export class MCPController implements Resource {
           description: server.description,
           source: server.source,
           authMode: server.authMode,
+          serviceHmacProfile: server.serviceHmacProfile,
           oauthTemplate: server.oauthTemplate,
           secretFields: server.secretFields,
           homepageUrl: server.homepageUrl,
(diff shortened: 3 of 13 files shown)


j4ys0n:
**Automated review** 🤖

**Summary of Changes**

Adds a closed `service_hmac` / `comms_management` MCP transport profile that signs outbound Streamable HTTP (and SSE fallback) requests with HMAC-SHA256 using a keyring loaded from a file path in `MSQ_COMMS_MCP_HMAC_KEY_FILE`. The change pins `@modelcontextprotocol/sdk` to `1.16.0`, introduces `ServiceHmacProvider` as a new process-level resource, threads it through `MCPController` and `MCPService`, and adds four new test suites covering golden vectors, reload, factory wiring, and end-to-end transport signing.

**Architecture & Organization**

- `src/services/serviceHmac.ts` is a self-contained 450-line module owning all keyring parsing, file loading, signing, and the `ServiceHmacProvider` lifecycle. The separation from `mcp.ts` is clean and the boundary is well-defined.
- `assertServiceHmacServerConfiguration` is called at four distinct call sites in `mcp.ts` (add, update, normalize, connect). This is intentional defense-in-depth but the repeated call in `connectUserToServer` (both the pre-check at line ~2529 and inside `_connectUserToServerInternal` at ~2566) means the same validation runs twice on the same object in the normal path—minor but worth noting.
- The `ServiceHmacServerValidationState` type in `mcp.ts` uses `unknown` for all fields, which is intentional to accept raw DB records, but it means the type provides no compile-time guidance for callers constructing the object.
- `createSseTransport` is exported for the first time in this PR solely to enable the transport test. If the SSE fallback path is considered internal, this export widens the public surface unnecessarily.

**Key Changes & Positives**

- 🟢 **Closed-profile enforcement is thorough**: `assertServiceHmacServerConfiguration` rejects every forbidden combination (external source, stdio, alternate URL, query/fragment, static headers, session ID, OAuth fields, user secrets) and is tested with a `test.each` matrix covering all 18 cases.
- 🟢 **Atomic keyring reload via serial promise queue** (`reloadQueue = reloadQueue.then(...)`) prevents concurrent file reads and ensures the in-memory keyring is replaced atomically.
- 🟢 **`redirect: 'error'`** is set unconditionally on every outbound signed request, preventing silent redirect-based SSRF against the internal endpoint.
- 🟢 **`O_NOFOLLOW`** on `open()` prevents symlink traversal at the OS level, and the test suite verifies this with an actual symlink.
- 🟢 **`authMode === 'oauth2' ? 'oauth2' : 'none'`** narrowing in `upsertInstall` calls correctly prevents `service_hmac` from leaking into the install record's auth mode.
- 🟢 **Golden vector tests** in `mcp-service-hmac.spec.ts` pin exact base64url signatures for POST, GET, and DELETE, making signing-string regressions immediately detectable.
- 🟢 **SSRF bypass for the internal URL is explicit and intentional**: `validateExternalMcpUrl` is skipped only after `isServiceHmacServer` confirms the exact pinned URL, and the existing external-URL validation path is preserved for all other servers.
- 🟢 **SDK pinned to exact version `1.16.0`** (not a range) after confirming `fetch?: FetchLike` is present in both transport option types—this prevents silent breakage from a future minor bump.

1. **Issue / Risk:** `reloadCandidate` in `serviceHmac.ts` swallows all errors with a bare `catch {}` block (lines ~420–435). A transient I/O error (e.g., EINTR, NFS hiccup) during a SIGHUP reload will silently disable signing until the next successful reload, with only a `warn`-level log and no structured error detail.
   - **Impact:** Operators have no way to distinguish a bad keyring from a transient I/O failure; the profile stays unavailable until the next SIGHUP or restart.
   - **Recommendation:** Log the caught error message (not the full stack) at `warn` level: `log({ level: 'warn', msg: \`Service-HMAC reload failed: ${err instanceof Error ? err.message : String(err)}\` })`. Avoid logging the error object itself to prevent accidental key-material leakage from future error types.
   - **Status:** 🟡 Needs review

2. **Issue / Risk:** `assertServiceHmacReady` in `mcp.ts` (line ~1817) throws `McpServiceHmacUnavailableError` when `this.serviceHmacProvider` is `undefined`, but `serviceHmacProvider` is typed as optional in `MCPService`'s constructor. In `index.ts` it is always constructed and passed, but the optional typing means a future caller could omit it and get a 503 instead of a programming error.
   - **Impact:** Silent misconfiguration at the call site; the error message ("profile comms_management is unavailable") is misleading when the real cause is a missing provider reference.
   - **Recommendation:** Make `serviceHmacProvider` required in `MCPService`'s constructor (it is always provided in `index.ts`) or add a distinct `McpValidationError` for the "provider not wired" case.
   - **Status:** 🟡 Needs review

3. **Issue / Risk:** The `serviceHmacTransportOptions` variable in `_connectUserToServerInternal` (lines ~2615–2716) is captured before the first `createTransport` call and reused for the SSE fallback. If `buildTransportOptions` is ever made async and the keyring rotates between the primary and fallback attempts, the fallback will use a signed fetch closure bound to the old key. Currently this is safe because `createSignedFetch` reads the keyring at call time (not closure time), but the comment in the code does not make this invariant explicit.
   - **Impact:** Low risk today; becomes a subtle bug if `createSignedFetch` is ever changed to capture the key at closure creation.
   - **Recommendation:** Add a brief inline comment at the `serviceHmacTransportOptions` capture site noting that `createSignedFetch` reads the active key lazily at each invocation, so reuse across the fallback is safe.
   - **Status:** 🟡 Needs review

4. **Issue / Risk:** The URL validation block in `updateServer` was restructured (lines ~3667–3697 in `mcp.ts`). The original code called `getKnownProviderUrlError` and `validateExternalMcpUrl` inside the `streamable_http` URL-setting block for external servers. The new code moves these checks to after `updatedServer` is assembled, but the `new URL(url)` parse (which was previously only for platform servers) now runs for all servers including external ones before the SSRF check.
   - **Impact:** Minimal functional change, but the ordering means a malformed URL that would have been caught by `validateExternalMcpUrl` is now parsed by `new URL()` first—this is actually slightly safer, but the diff removes the `source === 'external'` guard around `getKnownProviderUrlError`, meaning it now runs for platform servers too (which have no known-provider URLs to reject). This is harmless but unintentional.
   - **Recommendation:** Confirm the `getKnownProviderUrlError` call at line ~3700 is intentionally applied to all sources, or re-add the `source === 'external'` guard.
   - **Status:** 🟡 Needs review

5. **Issue / Risk:** The integration test in `mcp-streamable-http.integration.spec.ts` writes a keyring file to `tmpdir()` but does not call `provider.init()` before constructing `clientTransport`—`init()` is called inside the `try` block after `clientTransport` is constructed. If `init()` fails, `provider.stop()` in `finally` will still be called on an uninitialized provider.
   - **Impact:** `stop()` guards with `if (this.initialized)` so this is safe, but the ordering means the SIGHUP listener is registered after the transport is constructed, which is fine for the test but slightly inconsistent with production usage.
   - **Status:** 🟢 Correct

**Language/Framework Checks**

- `src/services/serviceHmac.ts`: `Buffer.from(keyBase64url, 'base64url')` followed by `key.toString('base64url') !== keyBase64url` correctly rejects padded input and round-trips the decode—this is the right way to validate unpadded base64url in Node.js.
- `src/services/serviceHmac.ts` `parseRfc3339`: The manual calendar validation (month, day, hour, minute, second, offset) is correct but redundant with the subsequent `Date.parse(value)` call. The `Date.parse` path is the authoritative check; the manual validation adds defense against `Date.parse` accepting out-of-range values in some V8 versions. This is intentional and acceptable.
- `src/services/mcp.ts` line ~2715: `serviceHmacTransportOptions ?? await this.buildTransportOptions(server, username)` — the `??` operator correctly short-circuits only on `undefined`/`null`, so a `{}` result from `buildTransportOptions` would not be reused. Since `serviceHmacTransportOptions` is set only for `isServiceHmacServer`, this is correct.

**Security & Privacy**

- The `resolveRequestUrl` check in `serviceHmac.ts` validates `origin`, `pathname`, `username`, `password`, and `hash` against the pinned constants before any signing material is accessed. This prevents the signed fetch from being redirected or confused to a different endpoint even if the SDK passes a modified URL.
- `headers.set(...)` (not `append`) is used for all five `X-MSQ-*` headers, ensuring a caller cannot inject a second value for any auth header even if the SDK passes pre-populated headers.
- The `RESERVED_HEADER_PREFIX` check (`x-msq-`) is case-normalized via `name.toLowerCase()` before comparison, correctly catching mixed-case bypass attempts (verified by the `x-MsQ-sIgNaTuRe` test case).
- `serviceHmacProfile` is included in error responses (`McpServiceHmacUnavailableError`) and in `UserVisibleMcpServer`. The key-file path, key bytes, timestamp, nonce, and signature are never serialized into any response type—confirmed by inspection of all `toUserVisibleServer` and `toMcpErrorResponse` paths.
- The `secretFields: authMode === 'service_hmac' ? undefined : effectiveSecretFields ?? []` assignment in `addServer` ensures no secret-field scaffolding is persisted for the HMAC profile.

**Tests**

- `test/mcp-service-hmac.spec.ts`: The `reloadCandidate` error-swallowing behavior (issue #1 above) is not directly tested—there is no test that injects a throwing `loadSymmetricKeyringFile` to confirm the warn log fires and `available` is set to `false`. Consider adding a test with a provider whose `keyFilePath` points to a file that is deleted between `init()` and `reload()`.
- `test/mcp-streamable-http.integration.spec.ts`: The `EPERM` guard on `httpServer.listen` silently skips the entire integration test on restricted environments. This is acceptable for CI but should be documented in the test comment so future readers understand the skip condition.

**Approval Recommendation**

**Approve with caveats**

- [ ] Address issue #1: log the caught error message (not the object) in `reloadCandidate` so operators can distinguish transient I/O failures from invalid keyrings.
- [ ] Confirm or fix issue #4: verify whether `getKnownProviderUrlError` should apply to platform servers after the `updateServer` restructure, or re-add the `source === 'external'` guard.
- [ ] Consider issue #2: decide whether `serviceHmacProvider` should be required in `MCPService` to prevent a misleading 503 from a wiring omission.

